PR #85, opened 2013-03-14
~/.ruby-version sets default ruby only for home directory tree

tmatilai:
The "[Default Ruby](https://github.com/postmodern/chruby#default-ruby)" section in the Readme indicates that _~/.ruby-version_ would set the default ruby the same way as calling `chruby`. Actually it only sets it for the home directory and its subdirectories.

I'm not sure if this is a problem for anyone, but maybe the wording in the readme should be tweaked a bit? If a true default value is wanted then `chruby_auto` should try to find and use _~/.ruby-version_ as a fallback before resetting to system ruby.

### DIFF
--- a/share/chruby/auto.sh
+++ b/share/chruby/auto.sh
@@ -22,6 +22,9 @@ function chruby_auto() {
 		dir="${dir%/*}"
 	done
 
+	chruby_auto_use "$HOME"
+	[[ $? == 2 ]] || return $?
+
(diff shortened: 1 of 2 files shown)


havenwood:
Support for a `.ruby-version` file in `/` was added in https://github.com/postmodern/chruby/commit/2e991e4d224409732e2f2b356fbd01a233faaf05.

PR #132, opened 2013-04-26
Update chruby.rb

adeluccar:
Add standard "put source ~/.bashrc in your ~/.bash_profile" advice in homebrew recipe.
If user runs login shell (such as Terminal) and ~/.bash_profile is not sourcing ~/.bashrc (assuming user added ../..chruby.sh source in ~/.bashrc) then chruby will fail to run.


morenoh149:
on osx .bashrc isn't executed by default. Should the readme instruct the user to add those lines to .profile instead?

postmodern:
@morenoh149 sounds like a good idea. Insert a note regarding OSX.

Issue #270, opened 2014-06-03
System-wide configuration for zsh


bjoernalbers:
chruby won't get sourced system-wide under zsh when using the README's config sample:

``` bash
if [ -n "$BASH_VERSION" ] || [ -n "$ZSH_VERSION" ]; then
  source /usr/local/share/chruby/chruby.sh
  ...
fi
```

I just [learned](https://github.com/sstephenson/rbenv/wiki/Unix-shell-initialization) that zsh ignores `/etc/profile` in favor of `/etc/zshenv`, so the config instructions might need to be updates.
This worked for me on OS X:

``` bash
# /etc/profile
if [ -n "$BASH_VERSION" ]; then
  source /usr/local/share/chruby/chruby.sh
  ...
fi

# /etc/zshenv
if [ -d /usr/local/share/chruby ]; then
  source /usr/local/share/chruby/chruby.sh
  ...
fi
```

PR #271, opened 2014-06-03
Fix system-wide config instructions for zsh.

bjoernalbers:
...as already mentioned in issue #270.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,17 +120,29 @@ source /usr/local/share/chruby/chruby.sh
 ### System Wide
 
 If you wish to enable chruby system-wide, add the following to
-`/etc/profile.d/chruby.sh`:
+`/etc/profile` or `/etc/profile.d/chruby.sh`:
 
 ``` bash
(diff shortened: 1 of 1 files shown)


postmodern:
From the `/etc/profile`:

<pre>
# It's NOT a good idea to change this file unless you know what you
# are doing. It's much better to create a custom.sh shell script in
# /etc/profile.d/ to make custom changes to your environment, as this
# will prevent the need for merging in future updates.
</pre>

zsh appears to load `/etc/profile`, which also loads `/etc/profile.d/*.sh`.

RobertAudi:
> zsh appears to load /etc/profile, which also loads /etc/profile.d/*.sh

@postmodern This doesn't appear to be the case in OS X.

I even `grep`-ed `/etc` and there isn't a single mention of `profile.d`

```
> sudo find /etc -type f -exec grep -H 'profile\.d' {} +
```

As a side note: `/etc/zshenv` is readonly. Not sure about the potential side-effects of making it readwrite for root.

PR #277, opened 2014-07-10
sort rubies versions properly

grimm26:
Having rbx-2.2.9 and rbx-2.2.10 in rubies and specifying "chruby rbx" would choose rbx-2.2.9 because chruby.sh just globs the contents of the rubies directories and so the members of the RUBIES array are sorted in ascii order: rbx-2.2.9 > rbx-2.2.10.

Using ls -v instead sorts the ruby versions properly.


grimm26:
closing in deference to #278 since more development has been done.

PR #316, opened 2015-01-27
Allow chruby-exec to set custom RUBIES

steakknife:
`chruby-exec` always sources `chruby.sh`, which nukes `RUBIES` with `RUBIES=()`

This PR add an ability to set RUBIES as another pre-argument for `chruby-exec`

(Tested, works.)


steakknife:
Squashed intermediate commits.

PR #319, opened 2015-01-31
Update test ruby to p481

tvon:
`2.0.0-p353` is no longer available on the rvm servers, which causes `test/setup` to fail.

http://rvm.io/binaries/osx/10.10/x86_64/


tvon:
Unnecessary with https://github.com/postmodern/chruby/commit/ef5e20789e8812cf988030d5cf1abcaefe6ef0b6

Issue #318, opened 2015-01-30
Version matching based on minor.patch over major.minor 


tvon:
chruby will match based on `${minor}.${patch}` instead of `${major}.${minor}`, e.g. specifying `2.0` will select `2.2.0` instead of `2.0.0`, and (for the sake of testing) specifying `1.5` will select `2.1.5`.  The latter would be a problem if `2.1.9` were to be released and the user only specified `1.9` in their `.ruby-version`.

```
tvon@roberto ~ % chruby --version
chruby: 0.3.9
tvon@roberto ~ % chruby
   jruby-1.7.16.1
   rbx-2.2.10
   ruby-1.9.3-p484
   ruby-1.9.3-p547
   ruby-2.0.0-p353
   ruby-2.0.0-p598
   ruby-2.1.3
   ruby-2.1.5
 * ruby-2.2.0
tvon@roberto ~ % chruby 1.9
tvon@roberto ~ % which ruby
/opt/rubies/ruby-1.9.3-p547/bin/ruby
tvon@roberto ~ % chruby 2.0
tvon@roberto ~ % which ruby
/opt/rubies/ruby-2.2.0/bin/ruby
tvon@roberto ~ % chruby 2.0.0
tvon@roberto ~ % which ruby
/opt/rubies/ruby-2.0.0-p598/bin/ruby
tvon@roberto ~ % chruby 1.3
tvon@roberto ~ % which ruby
/opt/rubies/ruby-2.1.3/bin/ruby
```

tvon:
[The issue](https://github.com/postmodern/chruby/blob/bbc43a0410a22b8ab0b1363fe672233e342c80f5/share/chruby/chruby.sh#L88):

```bash
case "${dir##*/}" in
	"$1")	match="$dir" && break ;;
	*"$1"*)	match="$dir" ;;
esac
```

So if you have `ruby-2.1.0` and `ruby-2.2.1` then `match` will match both, but end up being set as the last one it matches.  I don't know how to solve this with shell pattern matching, I think you'd want either `*-"$1"` or `^"$1"*` but not `*"$1"*`

PR #320, opened 2015-01-31
Adjust version matching

tvon:
This is based off of https://github.com/postmodern/chruby/pull/319

Addresses #318
- Breaks out ruby selection into function for testing.
- Removes `*"$1"*` matching and adds `ruby-"$1"*` and `"$1"*` matching.

For example, given the following:

``` bash
RUBIES=(
  /opt/rubies/ruby-2.0.0-p100
  /opt/rubies/ruby-2.1.1
  /opt/rubies/ruby-2.2.0
  /opt/rubies/ruby-2.2.1
  /opt/rubies/rbx-2.2.0
  /opt/rubies/rbx-2.2.1
)
```
- `chruby ruby-2.2` selects `/opt/rubies/ruby-2.2.1`
- `chruby rbx` selects to `/opt/rubies/rbx-2.2.1`
- `chruby 2.1` selects to `/opt/rubies/ruby-2.1.1`, this is the only change in the current behavior.  In master this would select `/opt/rubies/ruby-2.2.1`

### DIFF
--- a/benchmarks/chruby_auto.sh
+++ b/benchmarks/chruby_auto.sh
@@ -1,7 +1,7 @@
 root="${0%/*}/.."
 n=100
 
-RUBIES=("$root/test/opt/rubies/ruby-2.0.0-p353")
+RUBIES=("$root/test/opt/rubies/ruby-2.0.0-p481")
 
 . "$root/share/chruby/chruby.sh"
 . "$root/share/chruby/auto.sh"
--- a/benchmarks/chruby_use.sh
+++ b/benchmarks/chruby_use.sh
@@ -1,6 +1,6 @@
 root="${0%/*}/.."
 n=100
-ruby_dir="$root/test/opt/rubies/ruby-2.0.0-p353"
+ruby_dir="$root/test/opt/rubies/ruby-2.0.0-p481"
 
 . "$root/share/chruby/chruby.sh"
 
--- a/share/chruby/chruby.sh
+++ b/share/chruby/chruby.sh
@@ -58,6 +58,20 @@ EOF
 	fi
 }
 
+function chruby_ruby_select()
+{
+	local dir
+	for dir in "${RUBIES[@]}"; do
+		dir="${dir%%/}"
+		case "${dir##*/}" in
+			"$1") matched_ruby="$dir" && break;;
+			ruby-"$1"*) matched_ruby="$dir";;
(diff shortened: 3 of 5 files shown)


jrafanie:
Thank you for chruby/ruby-install. 

As a user, this PR is great. I love chruby and recommend it to my team but dislike the current behavior. What's needed to get this reviewed/merged?

I get burned by this all the time when I install new rubies. Currently chruby 2.1 selects 2.2.1 when I want 2.1.5.  Previously, chruby 2.0 selected 2.2.0 instead of 2.0.0. My brain just doesn't grok the current behavior.

I've considered removing rubies locally just so I don't have to deal with this problem but that seems silly.

tvon:
@postmodern anything I can do to improve this?

postmodern:
I'm a little hesitant about preferring `ruby-$1*` over just `-$1*`. See #278 which better sorts rubies by their version. I also experimented with [adding precedence to matching](https://github.com/postmodern/chruby/commit/3eba993c41d36590050b07fcbdc4f5b6327c8b66).

tvon:
My reasoning for `ruby-*` is that the only time people omit the implementation is when using MRI, at least so far as I've seen.  In other words, if there is only a version number specified then we're looking for `ruby-*`, if there is more specified (`rbx-2.2`) then the `"$1"*` match will catch it.  Unless I'm missing something, `*-$1` could match `2.1` to `rbx-2.1.1` which would be a mistake (I'm assuming even the Rubinius developers don't expect `2.1` to mean `rbx-2.1`).

Sorting would be an improvement either way as this PR does nothing to ensure that 2.1 picks the latest 2.1.x, but without other changes I don't think it would address 2.1 matching 2.2.1 instead of 2.1.x.

alindeman:
`chruby 2.1` selecting `2.2.1` was pretty confusing to me. I don't doubt there are some complicated issues around version preference, but this specific case was very surprising to me.

ccutrer:
@postmodern any update on this? @tvon's reasoning on matching `ruby-$1*` rather than `*-$1*` seems pretty convincing to me.

franklinyu:
@postmodern Any update on this? When typing `2.1` I also expect `ruby-2.1.1` (or the latest version of 2.1) while I would include the implementation if I am not using MRI, as stated by @tvon.

PR #328, opened 2015-03-21
support mruby with no `defined?` method

kbrock:
Hi,

`mruby` does not provide `defined?` and it looks like it won't mruby/mruby#1696
~~As of Mar 21, 2015, mruby-dev does no support `const_defined?` or `defined?`.~~

~~How about using `rescue` instead of `defined?` to determine `RUBY_ENGINE`.~~
**EDIT:** How about using `Object.const_defined?` to determine `RUBY_ENGINE`

Thanks
--K

### DIFF
--- a/share/chruby/chruby.sh
+++ b/share/chruby/chruby.sh
@@ -46,7 +46,7 @@ function chruby_use()
 	typeset unset RUBYGEMS_GEMDEPS
 
 	eval "$("$RUBY_ROOT/bin/ruby" - <<EOF
-puts "export RUBY_ENGINE=#{defined?(RUBY_ENGINE) ? RUBY_ENGINE : 'ruby'};"
+puts "export RUBY_ENGINE=#{Object.const_defined?(:RUBY_ENGINE) ? RUBY_ENGINE : 'ruby'};"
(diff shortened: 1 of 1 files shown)


cremno:
Um, but `Object.const_defined?(:RUBY_ENGINE)` works as expected?

brann:
https://github.com/mruby/mruby/issues/1696#issuecomment-34826066

In the referenced ticket, it is recommended to use const_defined? over defined?
rather than saying const_defined? does not work.

kbrock:
@cremno thanks. I was trying `Object.const_defined?(RUBY_ENGINE)`

This works for both mruby and mri 2.0.

Will reformat. this is great

cremno:
I see. But now you've introduced a typo (`RUBY_ENINE`).

kbrock:
Thanks @cremno

havenwood:
This looks good to me. Should it be `Module.const_defined?(:RUBY_ENGINE)` though?

cremno:
I prefer `Object` since `RUBY_ENGINE` (and global constants in general) is defined on it.

Issue #327, opened 2015-03-17
debian all, and jessie 


brann:
Firstly, for all debian, test/setup has:
```bash
 79         system_version="$(< /etc/debian_version | awk -F. '{print $1"."$2}')"
```
That  line does not do what is expected, on current wheezy or jessie. It puts an empty string into system_version.
 
This is probably what that line intends to be:
```bash
system_version="$(awk -F. '{print $1"."$2}' < /etc/debian_version)"
```
But also, it is used to build a url for
http://rvm.io/binaries/debian/

which only contains:
```
6/                                                 03-Aug-2013 09:48       -
6.0/                                               03-Aug-2013 09:48       -
6.0.4/                                             03-Aug-2013 09:48       -
7/                                                 08-Mar-2015 02:35       -
7.0/                                               08-Mar-2015 02:35       -
jessie_sid/   
```

So 7.8 for current wheezy which the corrected code generates, still breaks. A glance in 6 and 6.0.4 reveals it contains the same files with the same md5.  Perhaps there rvm.io should probably be maintaining more comprehensive lists. But you could probably get functional by going with the lead number.

So:
```bash
system_version="$(awk -F. '{print $1}' < /etc/debian_version)"
```
and finally, since they have no 8 there yet for jessie (8.0 is the current name for jessie test, which is the safer alternative to jessie unstable (jessie_sid which is there), setting jessie_sid for 8 with:

```bash
 78         system_name="debian"
 79         system_version="$(awk -F. '{print $1}' < /etc/debian_version )"
 80         if [[ $system_version == "8" ]]; then
 81           echo "setting debian jessie 8.0 to jessie_sid because 8.0 didn't exist at the time of writing this"
 82           system_version="jessie_sid"
 83         fi
```

Works around it for now.

If you have a preferred path forward let me know. Maybe if rvm.io aren't going to create a comprehensive list of version directories one solution could be to build up a list of candidate versions and try them iteratively if there's a 404 failure? 8.0,8,jessie_sid for example.

~~Finally after it does download i get errors on tests with shunit2 complaining about bell-style but they pass manually so I'm going to ignore that because I don't know what it's on about.~~ i had set bell-style none in my .bashrc but that should have only been done for login shells, or interactive shells. so my problem.

postmodern:
@mpapis ping

mpapis:
[...]

PR #331, opened 2015-03-24
debian wheezy/jessie use major version and pass tests

brann:
from issue #327


brann:
havenwood created links to 8 and 8.0 so $1.$2 works if you want to keep that.

the change in test/chruby_exec_test.sh is still required for passing zsh tests in debian.

Issue #337, opened 2015-06-22
Support envar to disable auto


sheldonh:
I can't submit a PR with tests for this, because `http://rvm.io/binaries/ubuntu/15.04` doesn't exist yet.

It would be great if `chruby_auto` in `auto.sh` had a guard clause that aborted if some environment variable was set, e.g. `RUBY_AUTO_DISABLE`.

At least on Ubuntu, chruby's setup helper arranges for `chruby.sh` and `auto.sh` to be loaded automatically from `/etc/profile.d/chruby.sh`.

It would be nice to be able to set an envar in my `.bashrc` that inhibits auto behaviour on such systems.

sheldonh:
@postmodern plans to tackle this from a different angle.

PR #338, opened 2015-06-22
Disable chruby_auto if RUBY_AUTO_DISABLE is set

sheldonh:
Fixes postmodern/chruby#337

### DIFF
--- a/share/chruby/auto.sh
+++ b/share/chruby/auto.sh
@@ -3,6 +3,8 @@ unset RUBY_AUTO_VERSION
 function chruby_auto() {
 	local dir="$PWD/" version
 
+	[ -z "$RUBY_AUTO_DISABLE" ] || return 0
+
 	until [[ -z "$dir" ]]; do
 		dir="${dir%/*}"
 
--- a/test/chruby_auto_test.sh
+++ b/test/chruby_auto_test.sh
@@ -121,6 +121,15 @@ function test_chruby_auto_invalid_ruby_version()
 		     "$expected_auto_version" "$RUBY_AUTO_VERSION"
 }
 
+function test_chruby_auto_disable()
+{
+	RUBY_AUTO_DISABLE=1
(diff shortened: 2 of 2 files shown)


postmodern:
Why can't you simply not load `auto.sh`, or is another program loading both by default? It's a separate file specifically to allow users to opt-out.

sheldonh:
I can edit the `/etc/profile.d/chruby.sh` that the setup helper writes. It's just root-owned, so I figured it would be nice for individual users to have a way to opt out.

postmodern:
@sheldonh which setup helper? Perhaps we could make it more configurable.

sheldonh:
Oh, sorry. I'm referring to the [setup.sh script](https://github.com/postmodern/chruby/blob/4572384143f3ee40c59a164635a0e773be3a58ac/scripts/setup.sh) referred to from the [README](https://github.com/postmodern/chruby/blob/26df12ca8741db54016ac4bffbd011e79cc3009c/README.md#setupsh).

postmodern:
@sheldonh ah hmm. Did someone else run the script and you don't have sudo privileges?

sheldonh:
Nope. I ran the script and I have root privilege. As the only user on my workstation, it's a non-issue. I can just edit the file (or not run it in the first place, sourcing just `chruby.sh` in my personal `.bashrc`).

My thinking went like this:

* The documentation advertises a chruby setup script.
* The chruby setup script activates `chruby_auto`, system-wide.
* This makes `chruby_auto` something of a default.
* Other people might expect that default.
* But I don't like it.
* So on multi-user systems, let unprivileged people who don't like it, disable it for themselves.

If you don't like the idea, you can bin it. I won't be offended. Chruby is a great little tool. Thanks.

postmodern:
I'm thinking of adding options to `setup.sh` to enable/disable things. `--no-auto` or maybe `--no-global`.

PR #339, opened 2015-07-09
Jruby fix

deivid-rodriguez:


deivid-rodriguez:
I've been using this for some days and it's definitely not a good fix... If you interrupt a command while the `eval` command is running, you'll end up in the Ruby directory... :-1: 

Not sure what a good solution to this is, but it's not this one... :(

deivid-rodriguez:
Closing this and opening an issue instead.

PR #345, opened 2015-08-25
chruby will now ignore lost+found directores in the rubies directories

ktgeek:


ktgeek:
I can't think of a way to make this smaller and not have shell glob errors.  It should work, though.

share/chruby/chruby.sh
```diff
@@ -2,7 +2,11 @@ CHRUBY_VERSION="0.3.9"
 RUBIES=()
 
 for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
-	[[ -d "$dir" && -n "$(ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
+    if [[ -d "$dir" && -n "$(ls -A "$dir")" ]]; then
+        for rdir in "$dir"/*; do
+            [[ -f "$rdir/bin/ruby" && -x "$rdir/bin/ruby" ]] && RUBIES+=("$rdir")
```
kbrock: I think you just need `[[ -x "$rdir/bin/ruby" ]] && RUBIES+=("$rdir")`

though to be honest, I did not test.

PR #347, opened 2015-09-07
Use Exception to catch LoadError

kbrock:
mruby does not define `LoadError` (nor `require`)

This works when I replace the `rescue LoadError` with `rescue`

thanks for this great library


postmodern:
I don't think we are officially supporting mruby, as it's missing many of the necessary constants, last time I checked.

kbrock:
@postmodern ok. If this change is too invasive, just close.

Thanks again for the tool

postmodern:
Wow, turns out mruby-1.2.0 was released three days ago. Just tested it, and it supports `RUBY_ENGINE` and `RUBY_VERSION`. Although, ruby-install would have to install mruby, and setup `ruby` and `irb` symlinks.

Rescuing Exception would prevent any potential errors from disrupting chruby's execution. It would also silently swallow any serious errors. Appears that mruby 1.2.0 raises a NoMethodError, due to require not being defined.

kbrock:
thanks again for the great tool.
closing

PR #348, opened 2015-09-07
trying out build with newer travis system

kbrock:
not sure if this will work since I removed the `rm rvm` lines.

We'll see if the `c language` has rvm installed in it.


kbrock:
looks like this has already been fixed.

Great gem - thanks

PR #351, opened 2015-11-18
Document COPR repository for Fedora users

nwallace:
I've created a COPR repository for chruby (and ruby-install). This adds installation instructions for Fedora users to use this repository

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,13 @@ chruby is included in the official [FreeBSD ports collection]:
 
     cd /usr/ports/devel/chruby/ && make install clean
 
+### Fedora
+
+Fedora users can install chruby from [this COPR repository]:
+
+    # dnf copr enable nwallace/ruby-tools
+    # dnf install chruby
+
 ### Rubies
 
 #### Manually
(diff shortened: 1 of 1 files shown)


postmodern:
Wow I didn't know about COPR! Is this Fedora's RPM equivalent of PPAs?

nwallace:
Yeah, very similar to PPAs or the AUR.  I think it was initially built as a build service for like continuous integration and stuff, but somebody made a yum plugin to download packages from there and now Fedora includes that plugin by default with DNF.

If you'd prefer the repo to be under your name for these two packages, that's totally fine with me. I didn't have to modify the RPM spec at all.  Might make the repo look a little more legit.  You'd probably want to set that up yourself, though

postmodern:
I wonder if there's a way to setup GitHub integration with it? Would be nice when we tag a new version or update the spec file, it kicks off a new build.

nwallace:
Yeah, that should be possible. COPR has an [API](https://copr.fedoraproject.org/api/) that supports initiating builds. You have to provide a URL to the SRPM (not the spec file), but we could probably tweak `make release` to generate one inside the `rpm` directory, then use the link to that file here on GitHub for the API call.

I'm brand new to Fedora, RPMs and packaging in general, so I don't really know what it's like outside of Fedora.  Fedora uses `rpmbuild` (among others) to build the RPMs. Can we assume the machine that runs the release has that package?  How do you build your RPMs?

postmodern:
I finally got around to setting up my own [Copr repository](https://copr.fedorainfracloud.org/coprs/postmodern/chruby/). Although, I'm getting mysterious build failures from my ruby-install repo.

nwallace:
When I was trying this stuff last year, there was a known issue in COPR where pretty much every time you create a new project, it's first build would fail. I don't remember the specifics (maybe it was something about key generation?), but once enough time has passed, just re-running the same build would work

PR #353, opened 2015-12-11
Change (( to [[ to support mksh (korn shell)

gregf:


postmodern:
Please see the [mksh](https://github.com/postmodern/chruby/tree/mksh) branch. I ended up hitting a dead-end when we I learned that `mksh` doesn't support `trap DEBUG`, so no auto-switching support.

postmodern:
Also, mksh supports `(( ))` arithmetic statements.

Issue #346, opened 2015-08-28
Is there a way to get chruby to automatically do something when switching?


tkelley65:
I have a ruby 1.8 install where I must set RUBYLIB manually before any ruby programs will work ... is it possible to have chruby do this automatically when switching to ruby 1.8? I don't see a way ... it would be nice to be able to take arbitrary actions when switching rubies ...

postmodern:
You would have to write your own post-switching commands system.

bronson:
Or use one of the zillions that already exist...  Here are some notes from a year ago when I needed to solve this.  (I ended up writing a one-off script, eschew overengineering)

* [autoenv](https://github.com/kennethreitz/autoenv): Overrides cd, that won't go well.
* [direnv](https://github.com/zimbatm/direnv): Well that's an awful lot of Go code.
* [anyenv](https://github.com/nikolay/anyenv): No README, looks kinda loose on security.
* [dotenv](https://github.com/bkeepers/dotenv): Ruby/Rails-specific, requires changing your source code.
* [chenv](https://github.com/betamatt/chenv): A lot of code plus a git-style UI.
  It's aopular name: [chenv](http://nanonanonano.net/linux/chenv), [chenv](https://github.com/bogdanjuszczak/chenv), etc.
* [aam](https://github.com/jlogsdon/aam): Apparently it's now named tem.
* [tem](https://github.com/Tapjoy/tem): Must be solving a completely different problem?

postmodern:
@bronson right, the real problem is Bash doesn't support something like `preexec_functions`. You have to either inject a command into `PROMPT_COMMAND` (only works in interactive mode) or use `trap DEBUG` (not inherited by sub-processes). So most of these Bash solutions clobber any previously inserted Bash "hooks".

ilikepi:
Would the idea proposed in #227 make this easier?

postmodern:
@ilikepi providing a function would allow users to manually set the `trap DEBUG` command. Manually injection into `trap DEBUG` is still impossibly/difficult unless `set -T` is set.

ilikepi:
Right... so does that mean you agree with this proposal (in a general sense)? I'm happy to put together a PR for further discussion.

bitboxer:
`rbenv` has [plugins](https://github.com/rbenv/rbenv/wiki/Authoring-plugins) that enable you to do things when switching versions and other things. Would love to see that in chruby, too.

Issue #227, opened 2013-12-02
Provide trap function for wrapping DEBUG signal traps


aprescott:
Occasionally `.bashrc` (or similar) has an existing `trap 'foo' DEBUG` which can conflict with `auto.sh`'s. I recently hit this problem since I noticed `.ruby-version` was being ignored.

The way I fixed this was to patch `auto.sh` to have:

``` sh
function chruby_trap() {
    [[ "$BASH_COMMAND" != "$PROMPT_COMMAND" ]] && chruby_auto
}

if [[ -n "$ZSH_VERSION" ]]; then
    if [[ ! "$preexec_functions" == *chruby_auto* ]]; then
        preexec_functions+=("chruby_auto")
    fi
elif [[ -n "$BASH_VERSION" ]]; then
    trap chruby_trap DEBUG
fi
```

Then in my `.bashrc`, after loading `chruby.sh` and `auto.sh`, after any other calls, I provide my own `trap`:

``` sh
# Reset color for command output and call this new `chruby_trap` function
trap 'foo_bar; chruby_trap' DEBUG
```

Obviously `chruby_auto` would also have worked, if I'd provided the explicit `$BASH_COMMAND != $PROMPT_COMMAND` check, but I figured it was possible that "setting up the trap" wouldn't necessarily always be the same as "call `chruby_auto`", hence the extra function.

I can submit a PR if you're happy with the idea (probably with refinements). Failing that, I can also submit a README note to mention that `auto.sh` will set a `DEBUG` trap and explain how to wrap it with `chruby_auto`.

ilikepi:
I happened upon this issue in the process of writing up a similar proposal.  I agree it would be beneficial for the call to chruby_auto to be wrapped in another function that could more easily be overridden.  I was going to propose a slightly different change however:

```diff
diff --git a/share/chruby/auto.sh b/share/chruby/auto.sh
index 88f3cb3..e3ad1ee 100644
--- a/share/chruby/auto.sh
+++ b/share/chruby/auto.sh
@@ -1,5 +1,9 @@
 unset RUBY_AUTO_VERSION
 
+function ruby_version_autoexec() {
+	chruby_auto
+}
+
 function chruby_auto() {
 	local dir="$PWD/" version
 
@@ -25,9 +29,9 @@ function chruby_auto() {
 }
 
 if [[ -n "$ZSH_VERSION" ]]; then
-	if [[ ! "$preexec_functions" == *chruby_auto* ]]; then
[...]

Issue #191, opened 2013-08-03
Invoke chruby_auto before the prompt prints in interactive mode


postmodern:
@zendeavor points out that it is confusing to users that `chruby_auto` is invoked before the command is executed and before the prompt is printed. If user's (ex: @jc00ke) include `$RUBY_ENGINE` and `$RUBY_VERSION` in their `$PS1`, then these values wont be updated until the command after the `cd` command is executed. @zendeavor recommends using different hooking strategies for interactive vs. non-interactive modes.
### Interactive
- bash: `PROMPT_COMMAND` 
- zsh: `precmd_functions`
- ksh: `PS1`
### Non-interactive
- bash: `trap DEBUG`
- zsh: `preexec_functions`
- ksh: `trap DEBUG`

This would invoke `chruby_auto` before `PS1` is rendered in interactive mode, but invoke `chruby_auto` before the command is even executed in non-interactive mode.

swrobel:
Thank you, @caleb for that workaround! Adding my +1 for seeing this added to chruby.

FranklinYu:
I wonder whether we need to detect interactive shell in `auto.sh`. As far as I understand, `auto.sh` is only intended for interactive shell; non-interactive shell should use `chruby-exec` which only uses `chruby.sh`.

PR #355, opened 2016-01-07
Wrap chruby_auto to ease augmentation (issue #227)

ilikepi:
This change implements the proposal in #227 and likely addresses #346 as well.  All tests pass on OS X 10.10.

It occurs to me that this is somewhat of a breaking change for zsh users who expect chruby_auto to be added directly to preexec_functions (for example, the code in [this comment](https://github.com/postmodern/chruby/issues/191#issuecomment-64091397) from issue #191).

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,9 @@ https://gist.github.com/1912050
 
 If you want to automatically run the version of a gem executable specified in 
 your project's Gemfile, try 
-[rubygems-bundler](https://github.com/mpapis/rubygems-bundler).
+[rubygems-bundler](https://github.com/mpapis/rubygems-bundler). If you want to
+modify how and when chruby_auto is called, you may override the
+`ruby_version_autoexec` shell function.
 
 ### Default Ruby
 
--- a/share/chruby/auto.sh
+++ b/share/chruby/auto.sh
@@ -24,10 +24,14 @@ function chruby_auto() {
 	fi
 }
 
+function ruby_version_autoexec() {
+	chruby_auto
+}
+
 if [[ -n "$ZSH_VERSION" ]]; then
-	if [[ ! "$preexec_functions" == *chruby_auto* ]]; then
-		preexec_functions+=("chruby_auto")
+	if [[ ! "$preexec_functions" == *ruby_version_autoexec* ]]; then
+		preexec_functions+=("ruby_version_autoexec")
 	fi
 elif [[ -n "$BASH_VERSION" ]]; then
-	trap '[[ "$BASH_COMMAND" != "$PROMPT_COMMAND" ]] && chruby_auto' DEBUG
+	trap '[[ "$BASH_COMMAND" != "$PROMPT_COMMAND" ]] && ruby_version_autoexec' DEBUG
 fi
--- a/test/chruby_auto_test.sh
+++ b/test/chruby_auto_test.sh
@@ -7,35 +7,35 @@ function setUp()
 	unset RUBY_AUTO_VERSION
 }
 
-function test_chruby_auto_loaded_in_zsh()
+function test_ruby_version_autoexec_loaded_in_zsh()
 {
 	[[ -n "$ZSH_VERSION" ]] || return
 
-	assertEquals "did not add chruby_auto to preexec_functions" \
-		     "chruby_auto" \
+	assertEquals "did not add ruby_version_autoexec to preexec_functions" \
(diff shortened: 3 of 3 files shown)


aprescott:
It's been so long that a lot of this has seeped out of my head by now, but I was just re-reading my comment at https://github.com/postmodern/chruby/issues/227#issue-23588607 and thinking about the difference between calling

```shell
chruby_auto
```

and

```shell
[[ "$BASH_COMMAND" != "$PROMPT_COMMAND" ]] && chruby_auto
```

As I understand it, this change would repeatedly call `chruby_auto` without the initial `"$BASH_COMMAND" != "$PROMPT_COMMAND"` check. Does that have any side-effect when using bash?

ilikepi:
@aprescott In my proposed change, the check of `$PROMPT_COMMAND` is still present within its original location in the DEBUG trap (line 36).  The check prevents the trap from executing when `$PROMPT_COMMAND` itself executes.  If the check weren't there and the user had `$PROMPT_COMMAND` set, the trap code would run twice:

```
$ export PROMPT_COMMAND='echo prompt'
prompt
$ pwd
/Users/ilikepi
prompt
$ function saytrap() {
>   echo trap
> }
prompt
$ trap 'saytrap' DEBUG
trap
prompt
$ pwd
trap
/Users/ilikepi
trap
prompt
```

I believe the check is most appropriate within the trap rather than in the wrapper function.  If the check were moved to within the wrapper, that could easily lead to issues when a user tried to override the wrapper.

HaleTom:
+1 for this PR.

I set my own `DEBUG` trap (to time each shell command), and it gets ugly if not all contained in one (well named) function.

Are there any objections to this? (I note it's over a year old)

postmodern:
How about providing a common `chruby_hook` function that provides an entry-point for any/all hooks that run after every command, which `auto.sh` adds to.

FranklinYu:
> How about providing a common `chruby_hook` function that provides an entry-point for any/all hooks that run after every command, which `auto.sh` adds to.

Isn’t that exactly `preexec`?

postmodern:
> Isn’t that exactly `preexec`?

Unfortunately Bash (in particular Bash 3 which macOS still ships) does not have `preexec`.

FranklinYu:
Then I’m once again advertising for https://github.com/rcaloras/bash-preexec, which saves us from manually fiddling with `trap` (which is difficult to get right, anyway). It also fixes #367 for free: https://github.com/rcaloras/bash-preexec/blob/master/test/bash-preexec.bats#L198-L212

FranklinYu:
For example, what if we have `chpython`, `chnode`, and `chperl`? Compare the hook solution

```bash
function chruby_hook() {
    chpython_auto
}

function chpython_hook() {
    chnode_auto
}

function chnode_hook() {
    chperl_auto
}
```

with the bash-preexec solution

```bash
preexec_functions+=(chruby_auto chpython_auto chnode_auto chperl_auto)
```

In addition, what about #399? Hence, my proposal: if preexec exist, use it (Bash or Zsh); fall back to `trap` (so that a fresh clean Bash “just works”).

postmodern:
> Then I’m once again advertising for https://github.com/rcaloras/bash-preexec,

How did I not see this sooner? [bash-preexec](https://github.com/rcaloras/bash-preexec) looks like it would solve our problems and allow for better interoperability with other tools that wish to hook `DEBUG`. I am a little hesitant about adding a dependency that's not readily available via package managers. However, we could vendor [bash-preexec](https://github.com/rcaloras/bash-preexec) and only load it when under bash?

ilikepi:
Presumably you'd also check to see if `bash-preexec` was already loaded before attempting to load the vendored version?

Personally I'm not sure the added cost of vendoring a dependency is a net gain over the proposal in this PR, or the alternate proposal in https://github.com/postmodern/chruby/pull/355#issuecomment-770474556.  Incorporating `bash-preexec` seems counter to the minimalist nature of chruby.  It could always be documented as a recommended addon for people who didn't want to mess with the DEBUG trap explicitly.

FranklinYu:
I’m recommending detecting `bash-preexec`. I’m not suggesting vendoring it.

ilikepi:
> I’m recommending detecting bash-preexec. I’m not suggesting vendoring it.

Then to me, that change seems like it would be independent of this PR.  Personally I'd prefer not to have to install a separate add-on in order to be able to more easily control when `chruby_auto` is called.

PR #359, opened 2016-02-17
System wide install: Ubuntu

### LINKED ISSUE #358: System wide Install: Ubuntu
Add the suggested snippet to `/etc/bash.bashrc`

taqtiqa-mark:
Closes #358.


postmodern:
Ubuntu still supports `/etc/profile.d/`. Although you may be invoking a non-interactive/non-login shell, which only loads the `*rc` config files.

PR #360, opened 2016-03-10
make tests pass on debian jessie

tomclegg:


postmodern:
Perhaps we should have `test/setup` exit with -1 if it could not download the ruby?

tomclegg:
That part (exit non-zero after an error) seems to work properly already. The trouble is just that (without this change) the _next_ time you run `make test`, the `test/opt/rubies` dir already exists, and according to Makefile that means it is not necessary to run `test/setup` any more. Therefore, tests just fail until you manually clean up and start over.

tomclegg:
btw I noticed a bit too late that the "just major version" change here duplicates #331. The discussion there says "8.0 is there now" but today 8.3 isn't there so it's broken again. I looked at the RVM source code (from which the code here was originally taken) and the current version uses only the major version. I figured things are most likely to keep working if chruby uses the same url RVM itself uses.

postmodern:
I'm going to experiment with trying to move the ruby version/path variables into the Makefile, that way we can define an explicit dependency on `path/to/test/ruby-X.Y.Z`.

tomclegg:
@postmodern if the idea is to avoid the extra "done" file, moving the "skip if done" responsibility from `Makefile` to `setup` (see amended commit) is probably easier...

postmodern:
@tomclegg ah yes, let's move the directory test to `test/setup`.

tomclegg:
@postmodern is 0b14dbc2 more to your liking?

PR #373, opened 2016-11-11
memoize ruby env on disk

### LINKED ISSUE #369: chruby.sh's use of eval is painfully slow with JRuby
I'm attempting to get our team of Rubyists to standardize on using chruby. So far everyone testing it has liked it a lot; except for the JRuby users. They reported painful slowness while evaluating chruby.

JRuby shines for web apps, but for CLI tasks the startup time is significant. I've traced chruby's slowness to `chruby.sh`'s `eval` block. That block launches an instance of JRuby to in order to ultimately set `GEM_HOME`, `GEM_PATH`, and `PATH`.

https://github.com/postmodern/chruby/blob/67b92499dfb804fde2b52567f93c7f366ac72d9d/share/chruby/chruby.sh#L46

Has anyone thought of an alternative to a shell `eval` running a full blown instance of JRuby in order to set those environment variables? We've thought of conditionally skipping over the `eval` and hoping that `bundle exec` helps fill in the gaps. And we've thought of working with the engine and version number specified in a project's `.ruby-version` file. But before we start down either of these paths, I figured I'd ask if anyone has thought of a way around `chruby.sh`'s `eval` block.

Any ideas?

steakknife:
Closes #369

### DIFF
--- a/share/chruby/chruby.sh
+++ b/share/chruby/chruby.sh
@@ -30,6 +30,22 @@ function chruby_reset()
 	hash -r
 }
 
+function chruby_generate_cached_env()
+{
+	"$SHELL" -c "unset JAVACMD JRUBY_OPTS; RUBYGEMS_GEMDEPS='' '$RUBY_ROOT/bin/ruby' -" <<'SCRIPT'
+env = {
+  :RUBY_ENGINE  => Object.const_defined?(:RUBY_ENGINE) ? RUBY_ENGINE : 'ruby',
+  :RUBY_VERSION => RUBY_VERSION,
+}
+begin
+  require 'rubygems'
+  env[:GEM_ROOT] = Gem.default_dir
+rescue LoadError
+end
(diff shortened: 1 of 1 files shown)


steakknife:
(The build on mac was already broken.)

share/chruby/chruby.sh
```diff
@@ -42,13 +58,12 @@ function chruby_use()
 	export RUBY_ROOT="$1"
 	export RUBYOPT="$2"
 	export PATH="$RUBY_ROOT/bin:$PATH"
+	local ruby_cached_env="$RUBY_ROOT/.ruby_env"
```
postmodern: You won't always have write access to the ruby's root.

nightscape: If this were made configurable by the user (opting out of it e.g. by not providing a `RUBY_ENV_CACHE`), could it be merged then?

Issue #364, opened 2016-04-14
.ruby-version not working for 2.1.10


HoneyryderChuck:
I have multiple 2.1.x versions installed, including 2.1.8 and 2.1.10 .

If my .ruby-version contains `2.1`, it automatically switches to 2.1.8, although I'd be expecting 2.1.10

deivid-rodriguez:
I'd say this is a dup of #363

Issue #363, opened 2016-04-07
Switching does not do what you expect with double digit patch level version numbers


deivid-rodriguez:
```shell
$ chruby
  jruby-1.7.24
  jruby-9.0.5.0
  ruby-2.0.0-p648
  ruby-2.1.10
  ruby-2.1.9
  ruby-2.2.4
 * ruby-2.3.0 
$ chruby 2.1
$ chruby
  jruby-1.7.24
  jruby-9.0.5.0
  ruby-2.0.0-p648
  ruby-2.1.10
 * ruby-2.1.9 
  ruby-2.2.4
  ruby-2.3.0
```

donv:
This is happening with JRuby now.

Issue #350, opened 2015-10-08
After upgrade from chruby-0.3.8 to chruby-0.3.9, chruby is not finding any installed rubies.


tllucero:
markov@wdc05:~/Downloads/chruby-0.3.9$ ./scripts/bug_report.sh
chruby: unknown Ruby: 2.2.3

## System

    Linux wdc05 3.19.0-30-generic #34-Ubuntu SMP Fri Oct 2 22:08:41 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
    GNU bash, version 4.3.30(1)-release (x86_64-pc-linux-gnu) [/bin/bash]
    chruby version 0.3.9 [/usr/local/bin/chruby-exec]

## Environment

    CHRUBY_VERSION=0.3.9
    SHELL=/bin/bash
    PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/usr/games:/usr/local/games
    HOME=/home/markov
    RUBIES=()
    RUBY_ROOT=
    RUBY_VERSION=
    RUBY_ENGINE=
    RUBY_AUTO_VERSION=
    RUBYLIB=
    RUBYOPT=
    RUBYPATH=
    RUBYSHELL=
    GEM_ROOT=
    GEM_HOME=
    GEM_PATH=

## Hooks

    trap -- '' SIGTSTP
    trap -- '' SIGTTIN
    trap -- '' SIGTTOU
    trap -- '[[ "$BASH_COMMAND" != "$PROMPT_COMMAND" ]] && chruby_auto' DEBUG

## Aliases

    alias 2015='cd /srv/Markets/EodData/Symbols/2015'
    alias __git_find_on_cmdline='__git_find_subcommand'
    alias alert='notify-send --urgency=low -i "$([ $? = 0 ] && echo terminal || echo error)" "$(history|tail -n1|sed -e '\''s/^\s*[0-9]\+\s*//;s/[;&|]\s*alert$//'\'')"'
    alias egrep='egrep --color=auto'
    alias fgrep='fgrep --color=auto'
    alias grep='grep --color=auto'
    alias jazz41='cd /opt/campbell/jazz41.com'
    alias l='ls -CF'
    alias la='ls -A'
    alias ll='ls -alF'
    alias ls='ls --color=auto'
    alias test01='. ~/commands/test01'
    alias tree='. ~/commands/tree.sh'
    alias tth='cd ~/programming/tth'
[...]

Issue #314, opened 2015-01-10
Wishlist: integrate with outside Ruby installations like Chef-DK


L2G:
"Embedded" Ruby environments like Vagrant and the Chef Development Kit have their own ideas for where things are supposed to be placed. Chef-DK in particular has its own preferred settings for `PATH`, `GEM_ROOT`, `GEM_HOME`, and `GEM_PATH` that at present don't play well with chruby's method of managing these variables.

postmodern:
We could potentially add a separate file that appends Chef's Ruby to `$RUBIES`. I'm not sure how we could handle custom `GEM_ROOT`, `GEM_HOME` without fundamentally breaking chruby. Also, why do you need to use chruby with Chef's embedded ruby?

L2G:
Because, for whatever reason, the makers of the `vagrant-berkshelf` gem insist they won't support its use outside of the Chef Development Kit. But I do not want to be tied to one version of Ruby (the MRI 2.1.4 interpreter that ships with Chef DK).

PR #374, opened 2016-11-11
RUBIES -> chruby_detect_all_rubies

steakknife:
Fix/supercede #278, #291, #314, #316, #318, #320, #345, #350, #363, #364

It's necessary to eliminate `RUBIES` environment variable to add/remove missing rubies in real-time and to save time. Keeping it around would be just pointless, technical debt tyranny.

### DIFF
--- a/share/chruby/chruby.sh
+++ b/share/chruby/chruby.sh
@@ -1,10 +1,55 @@
 CHRUBY_VERSION="0.3.9"
-RUBIES=()
 
-for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
-	[[ -d "$dir" && -n "$(ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
-done
-unset dir
+function chruby_list_rubies()
+{
+	local chruby_entry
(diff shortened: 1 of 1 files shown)


postmodern:
BTW, I eventually switched to your style of logic of re-searching the `rubies` directories each time.
https://github.com/postmodern/chruby/commit/c87f4cfee2f7c431be53fbf10a9be4f75664afdc

steakknife:
No big deal. I've been migrating things to use XDG conventions and work cleanly in contexts of non-root user, system, and vendored activation, similar to Python's virtualenv.

There is greater work to be done in shells and scripting:
- package management like zplug
- running faster by JITing
- loading commands as native shared objects
- simplification (zsh reminds me of Haskell and nix)
- avoiding unnecessary forking

postmodern:
@steakknife I'm super curious about XDG conventions and the shell. While XDG has been embraced by Linux GUI apps, it hasn't really been embraced by shell utils (that I've seen) and macOS users/developers still haven't adopted XDG.

Issue #367, opened 2016-08-18
Value of $_ overwritten in bash


HaleTom:
In `bash`, `$_` should be “the last argument of the last command executed”

Instead, this is overwritten:
```
> @ravi@boxy:~/.dotfiles(master* u+1)$ echo "This argument should be printed"
> This argument should be printed
> @ravi@boxy:~/.dotfiles(master* u+1)$ echo $_
> chruby_auto
> @ravi@boxy:~/.dotfiles(master* u+1)$ 
```

steakknife:
@HaleTom `trap` makes a lot of changes, if not cached correctly. Is it overwriting `$?` too?

HaleTom:
`$?` is fine, but `$#` changes...

HaleTom:
I raised this as a question on SO.

It's possible that this is a `bash` bug, see the `zsh` example working at: [bash: preserve $_ in a DEBUG trap](http://stackoverflow.com/q/40944532/5353461)

HaleTom:
Solution given on SO, I'll make a PR.

steakknife:
The SO answer seems to be incomplete as only the last argument is captured in `old_`.

```bash
bash-4.4$ f() { local old_="$1"; echo "=== $old_ ==="; : "$old_"; }; trap 'f "$_"' DEBUG
bash-4.4$ ls -lart >/dev/null
=== DEBUG ===
bash-4.4$ echo "$_"
=== -lart ===
-lart
bash-4.4$
```

Even `s/1/@/` doesn't help.

HaleTom:
@steakknife, the [`bash` manual](https://www.gnu.org/software/bash/manual/bash.html#Special-Parameters) says:

> expands to the last argument to the previous command, after expansion

So I assert that the SO answer is correct.

PR #381, opened 2016-12-03
Fix Issue #367: Value of $_ overwritten in bash

HaleTom:
Fix Issue #367 based on [this answer](http://stackoverflow.com/a/40944702/5353461)


postmodern:
I don't think how we hook in `chruby_auto` works within the unit-test functions. I had to explicitly invoke `chruby_auto` to simulate it running after other commands.

test/chruby_auto_test.sh
```diff
@@ -121,6 +121,12 @@ function test_chruby_auto_invalid_ruby_version()
 		     "$expected_auto_version" "$RUBY_AUTO_VERSION"
 }
 
+function test_chruby_auto_preserves_dollar_underscore()
+{
+	: expected
```
steakknife: Might want to be

```shell
: expected 1 2 3
assertEquals "value of $_ does not change" "$_" "expected 1 2 3"
```

HaleTom: It would be just `3` as it's the last argument only...

HaleTom:
@postmodern how should I deal with that in this PR?

HaleTom:
@postmodern Merry Festivus. Where to from here?

share/chruby/auto.sh
```diff
@@ -22,12 +23,13 @@ function chruby_auto() {
 		chruby_reset
 		unset RUBY_AUTO_VERSION
 	fi
+	: "$old_" # restore $_ (last argument of last command executed)
 }
 
 if [[ -n "$ZSH_VERSION" ]]; then
 	if [[ ! "$preexec_functions" == *chruby_auto* ]]; then
 		preexec_functions+=("chruby_auto")
```
postmodern: I believe zsh also supports the `$_` variable.

postmodern:
@HaleTom so `trap '...' DEBUG` does not capture commands invoked from within functions, only at the top level. This means `chruby_auto` is automatically invoked in your unit-test. We need to device a test that actually exercises this somehow.

<pre>
#!/bin/sh

trap 'echo "Command: $BASH_COMMAND"' DEBUG

func() {
	echo 'inside of a function'
}

echo 'outside of a function'
func
</pre>

share/chruby/auto.sh
HaleTom: It does, but it does not suffer from this  issue.

postmodern: I think it may be time to create a special `chruby_auto_bash` function to wrap the Bash-specific logic.

HaleTom:
`set -o functrace` is what you're after.

Adding this, the output is:

```
Command: echo 'outside of a function'
outside of a function
Command: func
Command: func
Command: echo 'inside of a function'
inside of a function
```

I'm not sure why `Command: func` is duplicated... let me know if it's an issue.

I'm having a brain-fart: how do I run the tests again? (I couldn't see anything in `CONTRIBUTING` about it).

Let me know if I can help out with anything from here.

share/chruby/auto.sh
HaleTom: FYI, the save/restore `$_` also works in `zsh`, even though it is not required as `zsh` saves/restores automagically.

HaleTom:
@postmodern Here is the code you want:

```
#!/bin/bash
set +o functrace
trap 'trap_func' DEBUG

trap_func() {
  [[ $BASH_COMMAND != "${FUNCNAME:0}" ]] && echo "Command: $BASH_COMMAND"
}

func() {
  echo 'inside of a function'
}

echo 'outside of a function'
func
```
Please let me know what the next step is.

HaleTom:
@postmodern, @steakknife, 
Any blockers to merge this?

HaleTom:
Just checking in on this as it's been one human gestation since the last comment.

What is the next step here?

test/chruby_auto_test.sh
HaleTom: The `:` is just a dummy command - one that returns true and has the side-effect of setting `$_`.

HaleTom:
I see that [Travis on Apple didn't seem to run correctly](https://travis-ci.org/github/postmodern/chruby/jobs/180919484) - it has null output.

Is this expected?  Any other blockers to merging?

test/chruby_auto_test.sh
postmodern: Could this be rewritten as explicit code that sets `$_` or returns an a return value?

postmodern:
@HaleTom Apple (_still_) ships with Bash 3, so unfortunately all code has to be Bash 3 compatible. I have since switched to GitHub Actions, however I still need to fix how the test suite downloads/installs the macOS test ruby, which must be compatible with GitHub Action version of macOS.

test/chruby_auto_test.sh
HaleTom: See ed69d75

HaleTom:
@postmodern I've made this mergable again.  I expect that accepting this as-is would be would be an improvement as-is.

postmodern:
Going to close this because GitHub can no longer rebase the branch (I definitely prefer rebasing vs merging `master`) and the `test_chruby_auto_preserves_dollar_underscore` test needs to explicitly call `chruby_auto`, as the `trap DEBUG` hook does not actually execute for commands within functions. Currently the `test_chruby_auto_preserves_dollar_underscore` test does not actually test whether `chruby_auto` preserves `$_`.

Feel free to `git rebase -i` the branch against the current `master` branch and squash the commits down and re-submit the PR. Otherwise, I might try to recreate the PR myself just to get the fix into `master` since preserving `$_` does seem like a valid bug fix. Sorry for taking so long to get back to this PR, as I've been working extra hard on other projects.

PR #384, opened 2016-12-16
Make chruby-exec exec

egh:
Previous, chruby exec would leave a parent bash process around. Now it will exec the ruby command, leaving no parent bash process.

This is important when, for instance, starting ruby processes from upstart. Without this fix, upstart will be tracking the parent bash process, not the child ruby process.

It is also, in my opinion, the correct thing to do. With this change, `chruby-exec` starts your ruby process and gets out of the way. Without this fix, you always have an old `bash` process hanging around.


postmodern:
Good points.

egh:
Thank you!

PR #397, opened 2017-11-07
Change options passed to ls on rubies search

CaptChrisD:
On macOS High Sierra the options to `ls` must be lower case.

Previously when running
`$ source /usr/local/share/chruby/chruby.sh`

would result in
`Unrecognized option: 'A'.`


havenwood:
The `-a` flag would stop ignoring entries that start with a dot. The desired behavior from the `-A` flag is to not list the implied `.` and `..` directories.

martinsp:
I have a macOS High Sierra and `ls -A` (with a capital 'a') is working correctly
```bash
$ sw_vers -productVersion
10.13.1
$ ls -A ~/Tmp/test
t1.txt t2.txt
```

PR #406, opened 2018-10-25
Use `command ls`, in case ls is an alias

benjamineskola:
This is perhaps an edgecase, but if ls is an alias/function it's possible that the output is not as expected, leading to failures.


havenwood:
This would resolve #413.

FranklinYu:
See also #424.

postmodern:
Manually reviewed and merged even though CI is currently broken.

PR #414, opened 2019-05-21
fix invald `ls -A` change to `ls -a`

benlieb:
Recently a lot of people (myself included) are using the lsd replacement of ls. It is common to alias ls=lsd.

When you do this lsd complains about the ls -A used here:

image

image

This PR simply replaces this with ls -a

### DIFF
--- a/share/chruby/chruby.sh
+++ b/share/chruby/chruby.sh
@@ -2,7 +2,7 @@ CHRUBY_VERSION="0.3.9"
 RUBIES=()
 
 for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
-	[[ -d "$dir" && -n "$(ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
+	[[ -d "$dir" && -n "$(ls -a "$dir")" ]] && RUBIES+=("$dir"/*)
 done
 unset dir
 


havenwood:
One option for you is to source chruby before you declare your lsd alias. That way, the system ls will work as expected.

This is an aside, but you might want to also checkout [exa](https://the.exa.website/), a modern replacement for ls.

havenwood:
Escaping the command with `\ls -A` or `"ls" -A` would avoid your alias. Using `command ls -A` would also override shell functions in addition to aliases. We are currently not escaping all commands.

benlieb:
Thanks for the info, and reply! I'll just put chruby before the alias.

share/chruby/chruby.sh
```diff
@@ -2,7 +2,7 @@ CHRUBY_VERSION="0.3.9"
 RUBIES=()
 
 for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
-	[[ -d "$dir" && -n "$(ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
```
ixti: You can simple quote `ls`:

``` bash
"$("ls" -A "$dir")"
```

Bash expands only unquoted aliases: https://www.gnu.org/software/bash/manual/html_node/Aliases.html#Aliases

Issue #422, opened 2019-11-01
A way to not set GEM_HOME if the installed ruby is under $HOME


eregon:
I would like a way to tell chruby to not set `$GEM_*` variables, as this is much safer (i.e., does not mix gems of different Ruby implementations) when e.g., developing TruffleRuby and executing two different Rubies without a `chruby` in between (see https://github.com/postmodern/chruby/pull/410#issuecomment-515233038).

Currently I'm doing this by using a [branch](https://github.com/postmodern/chruby/compare/master...eregon:do-no-set-gem-home) on my chruby fork, removing the code that sets `GEM_HOME`, but that's obviously not very convenient or maintainable, and I can't easily advise other TruffleRuby developers to do the same. I'd much rather this was possible in chruby itself.

Actually, there is already a way to do this, by running everything as `root`, but that's obviously not very safe for my use-case.
So I'm thinking to just extend the check `if (( UID != 0 )); then` to something like `if (( UID != 0 )) && [ "$CHRUBY_SET_GEM_HOME" != "false" ]; then`.

We could also automatically just detect if `$RUBY_ROOT` is under `$HOME` and not set `$GEM_*` vars in that case, but @postmodern had some concerns about that in https://github.com/postmodern/chruby/pull/410#issuecomment-515754740:
> Furthermore, there is value in keeping user-installed gems separate from the ruby, in case you need to delete/re-install one or the other.

@postmodern @havenwood What do you think?
I would like to have this available in chruby 1.0.0, I'm happy to make a PR.

eregon:
Proof-of-concept PR for `$CHRUBY_SET_GEM_HOME`: https://github.com/postmodern/chruby/pull/423

postmodern:
The reason why chruby sets `GEM_HOME` for non-root users is to protect the ruby's built-in gem home, which may not always be writable. This also helps keep the ruby's built-in gems from any user installed gems. Perhaps better logic would be to check if the ruby's `GEM_HOME` is writable? Another idea might be to split `chruby_use` into a lower-level function that only sets the ruby, which you could explicitly call instead of the main `chruby` function.

Also, for the record, I'm adverse to adding yet more `CHRUBY_` env variables to user's shells.

eregon:
> The reason why chruby sets `GEM_HOME` for non-root users is to protect the ruby's built-in gem home, which may not always be writable.

[...]

PR #423, opened 2019-11-01
Add an environment variable $CHRUBY_SET_GEM_HOME to disable setting $GEM_HOME

eregon:
This is a proof-of-concept PR to address #422.
It still needs documentation and a changelog entry.

@postmodern @havenwood What do you think?

### DIFF
--- a/share/chruby/chruby.sh
+++ b/share/chruby/chruby.sh
@@ -13,7 +13,7 @@ function chruby_reset()
 	PATH=":$PATH:"; PATH="${PATH//:$RUBY_ROOT\/bin:/:}"
 	[[ -n "$GEM_ROOT" ]] && PATH="${PATH//:$GEM_ROOT\/bin:/:}"
 
-	if (( UID != 0 )); then
+	if (( UID != 0 )) && [[ "$CHRUBY_SET_GEM_HOME" != "false" ]]; then
 		[[ -n "$GEM_HOME" ]] && PATH="${PATH//:$GEM_HOME\/bin:/:}"
 
 		GEM_PATH=":$GEM_PATH:"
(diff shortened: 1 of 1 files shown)


eregon:
@postmodern @havenwood Could we merge this? What do you think?

share/chruby/chruby.sh
```diff
@@ -13,7 +13,7 @@ function chruby_reset()
 	PATH=":$PATH:"; PATH="${PATH//:$RUBY_ROOT\/bin:/:}"
 	[[ -n "$GEM_ROOT" ]] && PATH="${PATH//:$GEM_ROOT\/bin:/:}"
 
-	if (( UID != 0 )); then
+	if (( UID != 0 )) && [ "$CHRUBY_SET_GEM_HOME" != "false" ]; then
```
havenwood: Other tests expressions use double brackets, which I think would be best to stick with for consistent compatibility and style.

My first impulse was to prefer `0` over `"false"` for falsey, but I don't have a strong defense.

eregon: OK, I'll switch to double brackets.

I think `false` is a bit more readable as in:
```bash
# In e.g., ~/.profile
export CHRUBY_SET_GEM_HOME=false

# Looks more cryptic to me:
export CHRUBY_SET_GEM_HOME=0
```

eregon: Done, please re-review.

havenwood:
This looks good to me.

The only question I would have is whether to check if the variable is set rather than `"false"` or `0`.  This seems like as good a solution as any to me, unless @postmodern has an opinion on it. It seems there are some shell libs (like [bash-boolean-helpers](https://github.com/Jeff-Russ/bash-boolean-helpers)) but hardly worth adopting here.

postmodern:
Posted some thoughts: https://github.com/postmodern/chruby/issues/422#issuecomment-557379476

I've seen people prefer testing for `"true"` instead of not `"false"`.

eregon:
> Posted some thoughts: [#422 (comment)](https://github.com/postmodern/chruby/issues/422#issuecomment-557379476)

Thanks, I replied there.

> I've seen people prefer testing for `"true"` instead of not `"false"`.

`CHRUBY_SET_GEM_HOME` has a default of `true`, so an explicit `false` seems the clearest to me.

eregon:
I made a PR which doesn't need user configuration and just does the "right thing" based on whether the default gem home is writable: https://github.com/postmodern/chruby/pull/431

It seems better than this PR, so I'll close this.

PR #424, opened 2019-11-13
Use bash globbing to set RUBIES

### LINKED ISSUE #413: alias for ls breaks chruby.sh
i recently aliased ls in .bashrc:
  alias ls='lsd'

at the bottom of .bashrc, i have
  source /usr/local/share/chruby/chruby.sh

which now errors out on the unsupported -A flag

a change of line https://github.com/postmodern/chruby/blob/4c4672214e2200271c3301a6f4e2cbb9445d6efd/share/chruby/chruby.sh#L5

from 'ls -A' to '/bin/ls -A' fixes this

ktchen14:
This is an alternative to #406. It fixes #413 and uses bash native functionality instead of `ls` at all. The impetus for this change is that an alias to `ls` breaks `chruby`. Specifically in my case I have:

```sh
alias ls=exa
```

Which doesn't have a `-A` flag.

### DIFF
--- a/share/chruby/chruby.sh
+++ b/share/chruby/chruby.sh
@@ -1,10 +1,23 @@
 CHRUBY_VERSION="0.3.9"
 RUBIES=()
 
-for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
-	[[ -d "$dir" && -n "$(ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
-done
-unset dir
+function chruby_rubies()
+{
+	if [ -n "$ZSH_VERSION" ]; then
+		setopt local_options nodotglob nullglob
+	else
+		trap "$(shopt -p dotglob nullglob)" RETURN
+		shopt -u dotglob
+		shopt -s nullglob
+	fi
+
+	local dir
+	for dir in "$@"; do
+		[[ -d "$dir" ]] && RUBIES+=("$dir"/*)
+	done
+}
+chruby_rubies "$PREFIX/opt/rubies" "$HOME/.rubies"
+unset -f chruby_rubies
 
 function chruby_reset()
 {


havenwood:
A shell-based solution would need to be portable to zsh since chruby support bash and zsh.

For the moment, you can source chruby before you set your exa alias for a fix.

share/chruby/chruby.sh
```diff
@@ -1,10 +1,14 @@
 CHRUBY_VERSION="0.3.9"
 RUBIES=()
 
+saveglob="$(shopt -p nullglob dotglob)"
```
havenwood: In zsh, the `shopt` equivalent would be `setopt`.

share/chruby/chruby.sh
```diff
@@ -1,10 +1,14 @@
...
+shopt -s nullglob
+shopt -u dotglob
 for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
-	[[ -d "$dir" && -n "$(ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
```
havenwood: Another thing to consider is the sort order. In zsh, you can get a proper numeric sort fairly easily, so `2.10` is greater than `2.9`.

Here is a zsh one-liner.
```sh
echo "${(@n)RUBIES}"
```
Or to print each Ruby in order:
```sh
printf "%s\n" "${(@n)RUBIES:t}"
```

havenwood:
See #278 for a stab at numeric sorting. We discussed the possibility of creating a shell library for portable numeric sorting, since it's an annoying gap. I'm for switching to a portable shell solution but think it should fix the numeric sorting issue too.

ktchen14:
@havenwood I just pushed a new version of the code which should be `zsh` compatible. I don't really use `zsh` but from a careful reading of the documentation this is the simplest version that I think works for both `bash` and `zsh`.

ktchen14:
Also this doesn't handle ordering. That looks hard to do in a portable fashion and I believe is out of scope for this PR.

ktchen14:
Ping. Please let me know if there are any changes I should make or if I should close this PR. Thanks!

Issue #390, opened 2017-07-12
update fedora copr repository


burningTyger:
F26 is out and I can't install with the f25 release. Thank you!

postmodern:
I've enabled `fedora-26-*` in the Copr Settings. Not sure if I also have to rebuild the packages to populate the new repos, or does Copr automatically handle that?

burningTyger:
I think you need to rebuild it. It doesn't install yet. Could you do the same for ruby-install? Thank you very much, I'm a happy user otherwise :)

postmodern:
The re-build apparently failed on the `make install` part. https://copr-be.cloud.fedoraproject.org/results/postmodern/chruby/fedora-25-x86_64/00584763-chruby/build.log.gz

Has the rpm spec changed since f25? Am I forgetting to set some install directive?

joerocklin:
Bump for fedora 27

FreedomBen:
@postmodern do you need/want some help with this?  I've never hacked on an rpm spec file but it would be super convenient to have this and ruby-install in the repo for f27

franklinyu:
Bump for Fedora 28.

taylorzr:
I'm a bit of a fedora noob, but I'm trying to use chruby on fedora 30, and getting the error:
```
$  sudo dnf copr enable postmodern/chruby
Error: This repository does not have any builds yet so you cannot enable it now.
```
Is this the same problem as this issue? 

Seems like chruby on fedora is broken? https://copr.fedorainfracloud.org/coprs/postmodern/chruby/ says build failed 2 years ago

FreedomBen:
@taylorzr I'm using brew for linux to install it now:  https://docs.brew.sh/Homebrew-on-Linux

`brew install ruby-install chruby`

PR #429, opened 2020-01-04
switch to an up2date copr

duritong:
I would maintain that copr


FranklinYu:
Would it be better to solve https://github.com/postmodern/chruby/issues/390#issuecomment-318863234 instead?

PR #433, opened 2020-02-27
document why not having shims is good

grosser:
I came here from https://github.com/rbenv/rbenv/issues/1223 :trollface:

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Changes the current Ruby.
 ## Anti-Features
 
 * Does not hook `cd`.
-* Does not install executable shims.
+* Does not install executable shims (avoids ~50ms overhead).
 * Does not require Rubies be installed into your home directory.
 * Does not automatically switch Rubies by default.
 * Does not require write-access to the Ruby directory in order to install gems.


peterb:
❤️

postmodern:
Decided to mention in 29af09c2642a7e8c4a894acaf84a3341c39623dd and 4f38caa621dc875533f345b3351ec80a52dd0a6e about when shims require constant regeneration, which is an annoying chore. I feel like rbenv could just optimize the shim regeneration or background the process and declare the problem solved.

eregon:
Those commits don't mention the performance cost of shims though, so it seems a separate concern.
:+1: for this PR.

peterb:
Had another look at this and I've changed my mind.

@postmodern I think the time it takes for the shim to execute is dependent on the system architecture and might involve binaries? If I'm right, then it feels to me like this is something that would be better off treated as an add-on, maybe by providing a hook for plug-ins.

@eregon It would be nice to include some real-world figures in the docs, but surely that would have to mention the system specifications? The laws of nature and mathematics would tell us that a single number isn't statistically meaningful.

PR #435, opened 2020-03-28
Fix tests

eregon:
There were still a few issues that prevented tests to pass on my local machine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ addons:
       - zsh
 
 install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install shunit2 bash; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then HOMEBREW_NO_AUTO_UPDATE=1 brew install shunit2 bash; fi
 
 before_script:
   - sudo rm -rf $HOME/.rvm $HOME/.rvmrc
--- a/Makefile
+++ b/Makefile
@@ -38,16 +38,16 @@ verify: $(PKG) $(SIG)
 	gpg --verify $(SIG) $(PKG)
 
 clean:
-	rm -rf test/opt/rubies
+	rm -rf test/fixtures/opt/rubies
 	rm -f $(PKG) $(SIG)
 
 check:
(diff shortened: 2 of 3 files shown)


eregon:
CI (for some reason TravisCI fails to use the commit status): https://travis-ci.org/github/postmodern/chruby/builds/668116142

postmodern:
Updated the `.travis.yml` format to use the new `addons: / packages: / homebrew:` syntax which apparently doesn't run `brew update` by default.

postmodern:
@eregon Grumble. Apparently there's an issue with TravisCI's macOS/XCode image and homebrew needing updating or upgrading to macOS > 10.12, but RVM only has binaries for macOS <= 10.12. Is avoiding a homebrew update really that necessary?

* https://travis-ci.community/t/macos-build-fails-because-of-homebrew-bundle-unknown-command/7296
* https://travis-ci.org/github/postmodern/chruby/jobs/669462555?utm_medium=notification&utm_source=email
* https://rvm.io/binaries/osx/

eregon:
> Is avoiding a homebrew update really that necessary?

It's not, feel free to revert that commit if it helps.
I thought it's kind of a waste of time to update so many packages but if it's actually helpful I guess why not.

I could move the CI to GitHub Actions if you're interested, which has a recent macOS runners. With `ruby/setup-ruby` we can also easily access a Ruby built on that platform without any RVM involved, and so we could simplify the test setup logic a lot. Thoughts?

eregon:
I tried to run the CI on GitHub Actions in https://github.com/eregon/chruby/commits/github-actions
But it seems there are issues to get an interactive terminal there: https://github.com/eregon/chruby/runs/560551417

postmodern:
Appears there's two separate bugs here. Travis CI has an issue with `test_chruby_auto_loaded_in_zsh` and `test_chruby_auto_loaded_twice_in_zsh` under both interactive and non-interactive.
https://travis-ci.org/github/postmodern/chruby/jobs/670947326#L5390

GitHub Actions lacks job control and somehow breaks a comparison deep within shunit2, in addition to complaining about no job control.
https://github.com/eregon/chruby/runs/560551417#step:7:20

PR #437, opened 2020-04-20
Update instructions for running on OS X

peterb:
On OS X .bash_profile has precedence over .bashrc, so asking users to source chruby in .bash_profile means that they'll have fewer issues. As a side-note, modifying files in /etc/ makes it a lot harder to track down problems with the environment if they should occur.


peterb:
😞 Closing this, will add another PR with [ci skip] in it.

PR #438, opened 2020-04-21
WIP: add more explanation around dotfiles on OS X to documentation. [ci skip]

peterb:
The experience of starting to use `chruby` on OS X is already really good compared to other systems. The idea is that this will make it a tiny bit better.💅

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ Add the following to the `~/.bashrc` or `~/.zshrc` file:
 source /usr/local/share/chruby/chruby.sh
 ```
 
-*Note:* OSX does not automatically execute `~/.bashrc`, instead try adding to `/etc/bashrc`.
+*Note:* OS X does not automatically execute `~/.bashrc`, instead try adding to `~/.bash_profile` or `/etc/bashrc`.
 
 ### System Wide
 
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ Add the following to the `~/.bashrc` or `~/.zshrc` file:
 source /usr/local/share/chruby/chruby.sh
 ```
 
-*Note:* OS X does not automatically execute `~/.bashrc`, instead try adding to `~/.bash_profile` or `/etc/bashrc`.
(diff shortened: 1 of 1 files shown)


postmodern:
The reason the documentation recommends `~/.bashrc` or `~/.zshrc`, instead of `~/.bash_profile` or `~/.zprofile`, is because the `profile` configure files are only loaded for interactive shell sessions. The `rc` configuration files are loaded for both interactive and non-interactive shell sessions, allowing chruby to be used in scripts, etc.

I would instead double check whether macOS's bash is still ignoring `~/.bashrc`.

peterb:
I'm sure that the `OS X 10.14` ignores `~/.bashrc`, not sure about `10.15`. I didn't know about the difference between the `*rc` and `*profile` files, will add some details to the description and continue looking into it.

FranklinYu:
This should almost certainly be added to `rc` instead of profiles. They serve very different purposes.

About macOS ignoring `~/.bashrc`, I don’t think that’s macOS specific. IIRC Bash just ignores `~/.bashrc` when it is the login shell, on all platforms. In Linux you can reproduce it (ignoring `~/.bashrc`) when you SSH into it; when you use X, then Bash won’t be your “login shell”.

Almost all Linux distribution pre-populates your `~/.bash_profile` with something like

```shell
if [ -t 0 ]
then source ~/.bashrc
fi
```

Finally if you are on macOS I would recommend Zsh, which is more up-to-date, and will be default since Catalina.

FranklinYu:
I think this in ReadMe suffices:

> Add the following to the `~/.bashrc` or `~/.zshrc` file:
>
> ```bash
> source /usr/local/share/chruby/chruby.sh
> ```

And I don’t agree with the next line:

> *Note:* OSX does not automatically execute `~/.bashrc`, instead try adding to `/etc/bashrc`.

I would consider creating a Pull Request to update that line to the correct instruction, as well as to explain why macOS *behaves as if* it ignores `~/.bashrc`. @postmodern Would you like that?

postmodern:
@FranklinYu yes I would prefer a note or a section dedicated to how macOS's bash differs from regular bash.

peterb:
@FranklinYu did you update this or shall I pick it up?

PR #439, opened 2020-05-04
chruby-exec: preserve working directory

egor-tensin:
Some versions of bash cd to $HOME for login shells (in /etc/profile).
It's kind of stupid and broken, but still, chruby should IMO make sure
that current working directory is preserved (especially since a myriad
other things can change it).


egor-tensin:
It's been so long, I switched to a different version manager.

PR #441, opened 2020-05-25
Force `ls` to be a command

FranklinYu:
This forces `ls` to be a command. Without `command`, `ls` function or alias might affect the output. In my case I alias `ls` to [`exa`](https://the.exa.website), which doesn’t support `-A` (actually I have no idea why `-A` was used in the first place).

### DIFF
--- a/share/chruby/chruby.sh
+++ b/share/chruby/chruby.sh
@@ -2,7 +2,7 @@ CHRUBY_VERSION="0.3.9"
 RUBIES=()
 
 for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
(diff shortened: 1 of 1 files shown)


FranklinYu:
I have no idea why the test fails for macOS; it doesn’t seem related to what I did here.

In addition test won’t run on my machine (macOS Catalina) since RVM no longer provides Ruby binary for latest macOS. As a result, there is no way for me to verify whether this (failing test case) is transient. Maybe manually trigger the test again?

FranklinYu:
I have no idea why I created this Pull Request when I already knows #406. Sorry.

Issue #447, opened 2020-10-02
"ruby" matches "truffleruby"


esotericpig:
With TruffleRuby installed, if you do `chruby ruby`, you'll get this:

```
   jruby-9.2.13.0
   jruby-9.2.9.0
   ruby-2.6.5
   ruby-2.7.0
   ruby-2.7.1
 * truffleruby-20.2.0
```

Which isn't expected I think. I expected to get `ruby-2.7.1`.

In `chruby()` in [chruby.sh](https://github.com/postmodern/chruby/blob/824db2625aa78d1bce2fd25c2fc9be76f161b332/share/chruby/chruby.sh#L90), I think it'd be good to add this line to match `ruby` first:

```bash
case "$ruby" in
	"$1")	match="$dir" && break ;;
	"$1"*)	match="$dir" && break ;; # NEW CODE
	*"$1"*)	match="$dir" ;;
esac
```

Is this fine? If so, I can make a pull request, or if someone can add this, that'd be awesome.

havenwood:
@esotericpig Yes, looks right to me. I think with the current matching, adding the line you propose is a good solution.

esotericpig:
I thought that it would be an easy fix, but actually, it will require parsing the version numbers like issue [#445](https://github.com/postmodern/chruby/issues/445).

In fact, these 2 issues could be merged into 1 (close this one or that one).

I'll leave it open for now for one of the maintainers to decide.

eregon:
Note: the rubies that `chruby` see could be any valid directory name, so it might be e.g. `foo-bar`, so there isn't always a version.
I'm not sure what's a good way to fix this.

FWIW in ruby/setup-ruby, we can reliably split ruby implementation and version, so then it's much easier to match e.g. `ruby` or `2.1` correctly.

esotericpig:
> Note: the rubies that `chruby` see could be any valid directory name, so it might be e.g. `foo-bar`, so there isn't always a version.
> I'm not sure what's a good way to fix this.
> 
> FWIW in ruby/setup-ruby, we can reliably split ruby implementation and version, so then it's much easier to match e.g. `ruby` or `2.1` correctly.

Well, in Ruby, it isn't so bad with regexes, but don't know how you can do it in Bash while keeping compatibility in mind for Zsh and all environments.

```Ruby
# Copied from https://semver.org
[...]

PR #448, opened 2020-10-15
Fix "chruby ruby" to match std Ruby first

esotericpig:
Fix `chruby ruby` to match the standard Ruby first over TruffleRuby/JRuby. Issue #447 

However, if there is only TruffleRuby/JRuby installed, `chruby ruby` should still match them.

### DIFF
--- a/share/chruby/chruby.sh
+++ b/share/chruby/chruby.sh
@@ -88,6 +88,7 @@ function chruby()
 				dir="${dir%%/}"; ruby="${dir##*/}"
 				case "$ruby" in
 					"$1")	match="$dir" && break ;;
+					"$1"*)	match="$dir" && break ;;
 					*"$1"*)	match="$dir" ;;
 				esac
 			done
--- a/test/chruby_test.sh
+++ b/test/chruby_test.sh
@@ -70,4 +70,31 @@ function test_chruby_invalid_ruby()
 	assertEquals "did not return 1" 1 $?
 }
 
+# "chruby ruby" should match "ruby" over "truffleruby"/"jruby"
+function test_chruby_match_ruby_first()
+{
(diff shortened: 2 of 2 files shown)


share/chruby/chruby.sh
```diff
@@ -88,6 +88,7 @@ function chruby()
 				dir="${dir%%/}"; ruby="${dir##*/}"
 				case "$ruby" in
 					"$1")	match="$dir" && break ;;
+					"$1"*)	match="$dir" && break ;;
```
eregon, line 91: Won't this pick the first `ruby-*` instead of the wanted last `ruby-*`, in case there are multiple like:
```
ruby-2.6.6
ruby-2.7.2
```

Could you add a test for that?

esotericpig: You're right. It doesn't work and grabs `ruby-2.6.6` first.

However, the current logic of this case statement requires not matching and falling to the end.

So if `truffleruby` is installed and do `chruby ruby`, it will grab `truffleruby` because it is the last one ('t' comes after 'r'). This is for the current logic (without my pull request).

So, we would need to change the entire logic to parse the version number and grab the latest version number. This is outside of my capability in bash, so I don't know what to do.

Maybe I can just delete my pull request, and someone more knowledgeable in bash can make these changes. It will require a rewrite in the logic around this case statement parsing the version numbers.

Anyway, here is the new test code I used:

```bash
function test_chruby_grab_latest_version()
{
	RUBIES=("/path/to/ruby-2.6.6" "/path/to/ruby-2.7.2")

	local result=$(chruby "ruby" 2>&1)

	echo "$result"
}
```

esotericpig:
The entire logic will require being changed to parse the version numbers.

Unfortunately, this is outside my ability in bash at the moment.

PR #466, opened 2021-10-08
Added Zsh configuration instruction for OSX

erkhem42:
Zsh configuration instruction for OSX wasn't accurate. Therefore, added
the related snippet

Note: 
I found actual installed directory by looking up source code (`opt_share` variable). So, I think it could help someone like me out there

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ source /usr/local/share/chruby/chruby.sh
 
 ``` bash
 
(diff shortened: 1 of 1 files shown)


postmodern:
Can we convert that to the default brew prefix directory? I'm worried that running `brew --prefix` on every shell startup will slow down opening a new shell.

PR #470, opened 2022-01-22
Use bash-preexec when it is available

FranklinYu:
(Note: this is a draft request because it is waiting for decisions in rcaloras/bash-preexec#123

### DIFF
--- a/share/chruby/auto.sh
+++ b/share/chruby/auto.sh
@@ -24,7 +24,7 @@ function chruby_auto() {
 	fi
 }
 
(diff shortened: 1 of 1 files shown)


postmodern:
This seems like something that needs a test to catch future breaking changes. Could add a test file that checks for bash (`[[ -n "$BASH_VERSION" ]]` or even `[[ ! -n "$ZSH_VERSION" ]]`?), source in `bash_preexec` before `chruby/auto.sh`, then check if `preexec_functions` was appended.

PR #471, opened 2022-01-28
Fix tests so that they also work locally

FranklinYu:
The test works in GitHub Action environment, but it doesn’t work on all environments. This PR is trying to make the test more robust, and allow it to run on more environments (such as the one I develop on).

### DIFF
--- a/test/unit/chruby_reset_test.sh
+++ b/test/unit/chruby_reset_test.sh
@@ -15,10 +15,13 @@ function test_chruby_reset_hash_table()
 		assertEquals "did not clear the path table" \
 			     "" "$(hash)"
 	elif [[ -n "$BASH_VERSION" ]]; then
+		export LANG=en_US.UTF-8
 		chruby_reset
 
 		assertEquals "did not clear the path table" \
 			     "hash: hash table empty" "$(hash)"
+
+		unset LANG
 	fi
 }
 
--- a/test/unit/chruby_reset_test.sh
+++ b/test/unit/chruby_reset_test.sh
@@ -15,13 +15,14 @@ function test_chruby_reset_hash_table()
 		assertEquals "did not clear the path table" \
 			     "" "$(hash)"
 	elif [[ -n "$BASH_VERSION" ]]; then
+		local old_lang="$LANG"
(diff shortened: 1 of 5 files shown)


test/unit/chruby_reset_test.sh
```diff
@@ -15,10 +15,13 @@ function test_chruby_reset_hash_table()
 		assertEquals "did not clear the path table" \
 			     "" "$(hash)"
 	elif [[ -n "$BASH_VERSION" ]]; then
+		export LANG=en_US.UTF-8
```
FranklinYu, line 18: Note: tested on my machine, setting system language to Chinese. Without this change, the error message contains Chinese.

FranklinYu:
Quote from Bash manual supporting last commit:

> If a numeric argument is given to `return`, that is the function’s return status; otherwise the function’s return status is the exit status of the last command executed before the `return`.

Example:

```bash
func() { return; }

false
func || echo 'Error!'
```

test/unit/chruby_reset_test.sh
postmodern, line 18: We should probably do this for `zsh` as well, yes?

test/unit/chruby_reset_test.sh
```diff
@@ -15,10 +15,13 @@ function test_chruby_reset_hash_table()
...
 		chruby_reset
 
 		assertEquals "did not clear the path table" \
 			     "hash: hash table empty" "$(hash)"
+
+		unset LANG
```
postmodern, line 24: `LANG` is usually already set, so many store the original `LANG` value and reset it?

postmodern (commented):
Minor question regarding `bash` vs `zsh` with `LANG`.

test/unit/chruby_reset_test.sh
FranklinYu: The Zsh test above only compares it with empty string, which doesn’t sound like language-specific.

test/unit/chruby_reset_test.sh
FranklinYu: Done.

postmodern:
Merged!

PR #475, opened 2022-03-04
Recommend `setup.sh` as the default installation method.

ragaskar:
This would've saved me a few minutes of wondering why `chruby` was "command not found" after compiling it on my system. 😄 

I mean, my fault for not fully reading the instructions, but I suspect that's how many of us operate. 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,13 +37,13 @@ Changes the current Ruby.
 
 ## Install
 
-```shell
-wget -O chruby-0.3.9.tar.gz https://github.com/postmodern/chruby/archive/v0.3.9.tar.gz
-tar -xzvf chruby-0.3.9.tar.gz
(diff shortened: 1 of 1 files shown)


postmodern:
The reason why `make install` does not also install configuration is I don't want to choose for the user how chruby is configured/loaded. Users can choose whether chruby is loaded via `/etc/profile.d/` or in `~/.bashrc`/`~/.zshrc`.

ragaskar:
Yes, that makes sense to me. Sorry if the PR wording is unclear: I'm proposing a change that would guide most users towards running setup.sh vs running `make install`. I suspect folks who want to customize how chruby is installed will read the whole doc, whereas users who just want a `chruby` asap would be better served by just running the ./setup.sh script.

PR #492, opened 2023-04-26
[WIP] remove call to external ls when loading chruby

kbrock:
### before

```
/Users/kbrock/dotfiles/bashrc.d/chruby.sh

real	0m0.004s
user	0m0.001s
sys	0m0.001s
```

### after

```
/Users/kbrock/dotfiles/bashrc.d/chruby.sh

real	0m0.000s
user	0m0.000s
sys	0m0.000s
```

### why?

I was trying to figure out ways to make nvm.sh load faster on my system.
chruby loads so much faster, and when I looked to see why, I was distracted by the call to `ls -A "$dir"`. It was using a builtin to list the files in the directory, but an external command to detect if there were files in the directory.

So yea, I converted the already quick method to made it even faster.

- I tested with my files in `~/.rubies`, and a non-existent `/opt/rubies`.
- I did not test this with a directory with a space in the name, but didn't see how that could be possible other than having a `PREFIX` with a space in the name.
- I did not test with rubies that start with a period, but it looked like the original script did not handle that case either. And it didn't seem valid.

This is yak shaving, and you probably can think up a better implementation. But it seems to work in `zsh` and `bash`.

If you prefer not to introduce change (because change always introduces risk), then that is cool. Just close.
I just saw this change and wanted to share.

As you will note, `nvm` is noticeably slower and the reason for my original tangent:

```
/Users/kbrock/dotfiles/bashrc.d/node.bash

real	0m0.177s
user	0m0.033s
sys	0m0.089s
```

Thanks again for such a great tool. And it already performs heads and shoulders better than other tools out there. (ruby or other)

refs:
- https://github.com/kbrock/dotfiles/blob/master/bashrc.d/chruby.sh
- https://github.com/kbrock/dotfiles/blob/master/bashrc.d/node.bash (since changed to speed this up)

### DIFF
--- a/share/chruby/chruby.sh
+++ b/share/chruby/chruby.sh
@@ -2,9 +2,10 @@ CHRUBY_VERSION="0.3.9"
 RUBIES=()
 
 for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
-	[[ -d "$dir" && -n "$(command ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
+  [[ -d "$dir" ]] && dir_contents=("$dir"/*) && \
+    [[ ${#dir_contents[@]} -gt 0 ]] && RUBIES+=(${dir_contents[@]})
 done
-unset dir
+unset dir dir_contents
 
 function chruby_reset()
 {


kbrock:
WIP: sorry, in zsh it worked for non-existing director, or a directory with rubies, but not for a directory that was empty

and zsh is complicated due to the default of blowing up when there is a bad glob

running `setopt nonomatch` will get around this, but this started to get too complicated.
Since it is sourced, modifying that value in the parent shell is not so good.

postmodern:
@kbrock yep, we ended up using `ls -A` to filter out non-existent directories or empty directories, and to behave consistently even if the user has enabled some weird glob-option mode.

In the [1.0.0](https://github.com/postmodern/chruby/tree/1.0.0) branch, I changed `chruby` to [lazy enumerate the directories on each call](https://github.com/postmodern/chruby/blob/52dae1987a860415a4dbdbe9ff22c9b4ba0389d6/share/chruby/chruby.sh#L4-L32) with an extra directory existence check which should filter out non-existent or empty directories.

kbrock:
ooh, sweet!

Ugh, I guess I have been living in the homebrew world and forgot about version 1.0.

postmodern:
I've been slowly working on 1.0.0 branch where I can make breaking changes. The last feature I was trying to work on is adding a `auto.sh` like feature but for setting up `alias`es for `bundle exec` commands; had trouble finding all executables from bundler's API. I also plan on eventually releasing a 0.4.0 which adds some additional features and relieves some of the pain some users are feeling, such as the slow load times.

kbrock:
1. Just to be clear, 3.9 is plenty fast. This all started for me because `nvm` is slow and I was looking at `chruby` as an example of how to make `nvm` as quick as `chruby`.
2. Most people I know have a `be` or `ber`/`beer` (`bundle exec` and `bundle exec ruby`). I went a little too far and have a "smart" `beer` command that calls `ruby`, `rails`, `rake`, or `rspec` depending upon the parameters. Since rails introduce something similar to their `rails` command, it seems they felt a similar pain. Wonder if the added lines of code justify the additional functionality in `chruby`?

PR #493, opened 2023-04-26
move ruby discovery code into a function

kbrock:
When you install a new ruby in a different terminal shell, chruby does not pick up on the new ruby version.

The solution is to to source chruby.sh again.
Introducing chruby_discover which will allow any shell to pick up on the new ruby versions

---

Yes, I was trying to "optimize" this functionality in #492 - but this PR is not an optimization play.

I have had this function defined in my bashrc for a while and thought others may want it as well.

### DIFF
--- a/share/chruby/chruby.sh
+++ b/share/chruby/chruby.sh
@@ -1,10 +1,13 @@
 CHRUBY_VERSION="0.3.9"
-RUBIES=()
 
-for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
-	[[ -d "$dir" && -n "$(command ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
-done
-unset dir
+function chruby_discover()
+{
+	RUBIES=()
+	local dir
+	for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
+		[[ -d "$dir" && -n "$(command ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
+	done
+}
 
 function chruby_reset()
 {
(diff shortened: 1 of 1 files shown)


share/chruby/chruby.sh
```diff
@@ -1,10 +1,15 @@
...
 
-for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
-	[[ -d "$dir" && -n "$(command ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
-done
-unset dir
+function chruby_discover()
+{
+  RUBIES=()
+  local dir
+	for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
```
postmodern, line 7: Shell script must be hard-tab indented.

share/chruby/chruby.sh
```diff
@@ -1,10 +1,15 @@
...
+		[[ -d "$dir" && -n "$(command ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
+	done
+}
+
+chruby_discover
```
postmodern: I tend to put any non-function code at the end of the file.

postmodern (changes requested):
Spotted a few stylistic things.

postmodern:
Since you're adding a new function, I would technically need to merge this into the [0.4.0](https://github.com/postmodern/chruby/tree/0.4.0) branch.

kbrock:
Fixed. Thanks for the feedback. I forgot all about tabs.

It is ok if you don't want to go this route and bump the major version.
If you like this route, do you need me to add documentation or changelog or anything?

PR #503, opened 2025-01-10
Fix RPM build on COPR

gschlager:
I wanted to build chruby in my Fedora COPR at https://copr.fedorainfracloud.org/coprs/gschlager/ruby/ and noticed that it didn't work anymore. I had to make changes to `rpm/chruby.spec` and delete `rpm/sources` to make the build work.

This is related to https://github.com/postmodern/ruby-install/pull/492


postmodern:
I'm pretty sure we need something like `sources` to check the integrity of the downloaded source archive. Will definitely accept the `BuildRequires` change.

